PR #4, opened 2019-08-02
Finished Problem1.

Shazmaan:
Will finish problem2 after class in order to have a better understanding before writing code.

### DIFF
--- a/Problem1.cpp
+++ b/Problem1.cpp
@@ -1,0 +1,59 @@
+//
+// Created by shazm on 8/1/2019.
+//
+
+#include <iostream>
+#include <algorithm>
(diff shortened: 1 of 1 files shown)


Problem1.cpp
```diff
@@ -0,0 +1,59 @@
...
+        return image;
+    }
+private:
+    bool validate(int i, int j, vector<vector<int>>& image){
+        if(i>=0 && i<image.size() && j<image[0].size() && j>=0){
+            return true;
+        }
+        return false;
+    }
+};
```
super30admin: Thats a good Solution as well.

PR #5, opened 2019-08-02
Completed DFS-1 - Problem 1

bharatreddypothireddy:


Problem1.java
```diff
@@ -0,0 +1,45 @@
...
+            }
+            if(j!=image[0].length-1 && image[i][j+1] == old){
+                image[i][j+1] = newColor;
+                queue.add(new int[]{i,j+1});
+            }
+        }
+        return image;
+    }
+}
+
```
super30admin: Thats a  nice solution .. Also try the same using DFS for practice.

PR #48, opened 2019-12-27
Completed DFS 1

RishikaMachina:


Problem_2.py
```diff
@@ -0,0 +1,36 @@
+# Runs on Leetcode
+# Runtime - O(n)
+# Memory - O(1)
+
```
neha79091: Space complexity should be O(n)

RishikaMachina: True

neha79091: good

PR #53, opened 2020-01-24
Flood Fill & 01Matrix

BhushanVasisht:


neha79091 (commented):
Good Work Bhushan

Problem2.java
```diff
@@ -0,0 +1,71 @@
+// Time Complexity :
+//      n is the number of nodes in tree
+//      updateMatrix() - O(n)
+//
```
neha79091: you should write n as number of elements in matrix as there is no tree here

Problem2.java
```diff
@@ -0,0 +1,71 @@
...
+//      n is the number of nodes in tree
+//      updateMatrix() - O(n)
+//      
+// Space Complexity :
+//      updateMatrix() - O(n)
+//
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+// BFS
+
```
neha79091: Write DFS solution too

PR #103, opened 2020-05-21
Completed DFS-1: Please Review

anikpait:


01Matrix.java
```diff
@@ -0,0 +1,59 @@
...
+        for(int[] dir: dirs){
+            int r = i1 + dir[0];
+            int c = j1 + dir[1];
+            Pair p = new Pair(r,c);
+            if(!visited.contains(p)){                                                               // if the pair has not been visited yet
+            if(r >=0 && c >=0 && r<matrix.length && c <matrix[0].length && matrix[r][c]==-1){           // If the neighbor is a -1
+                matrix[r][c] = matrix[i1][j1] + 1;                                          // Add it to the distance since its a non 0 and update the distance of the new cell where I am reaching
+                q.add(p);
+                visited.add(p);                                                         // Marked the cell visited and add it to the queue
+            }
```
anikpait, line 54: Hi @jaspindersingh83 ,
The code is working for matrix[r][c] = -1 because I am mutating the array. But when I am trying by creating HashSet Visited here and checking if Visited contains the pair, it is not working correctly because the existing pairs in visited and the new pairs formed in r, c (that might be in visited) but it is treating them as new pairs whenever I am doing
Pair p = new Pair(r,c) though they might be creating the already existing pairs in visited.
Is it possible? Hence, visited should always be a matrix?

01Matrix.java
```diff
@@ -0,0 +1,59 @@
...
+        while(!q.isEmpty()){
+        Pair fetch = q.poll();                                                  // Fetch the pairs
+        int i1 = fetch.getRow();
+        int j1 = fetch.getCol();
+        for(int[] dir: dirs){
+            int r = i1 + dir[0];
+            int c = j1 + dir[1];
+            Pair p = new Pair(r,c);
+            if(!visited.contains(p)){                                                               // if the pair has not been visited yet
+            if(r >=0 && c >=0 && r<matrix.length && c <matrix[0].length && matrix[r][c]==-1){           // If the neighbor is a -1
```
anikpait, line 50: @super30admin ,
The code is working for matrix[r][c] = -1 because I am mutating the array. But when I am trying by creating HashSet Visited here and checking if Visited contains the pair, it is not working correctly because the existing pairs in visited and the new pairs formed in r, c (that might be in visited) but it is treating them as new pairs whenever I am doing
Pair p = new Pair(r,c) though they might be creating the already existing pairs in visited.
Is it possible?

anikpait: @sowmyadvn

iivaishnavii: Hi Anik, as discussed, two objects with the same attributes are still different. Hence, using a hashset<pairs> will not work..Like you mentioned, just mutate the array (since we know the array is just 1's and 0's) to keep track of visited elements. Maintaining a visited matrix would be an overkill for the problem too.

anikpait: Yes, thanks!

PR #247, opened 2020-11-12
Completed both the problems

kateganesh:


samarthjj:
Good work Ganesh, especially with the description of the approach.

samarthjj:
Good work Ganesh, especially with the description of the approach.

PR #329, opened 2021-01-26
Completed :Please review

akuma0639:


maithrreye:
Good Job ! Please add TC and SC for both the problems

maithrreye:
Good Job ! Please add TC and SC for both the problems

PR #496, opened 2021-08-31
done pls review

VishLivecoder:


Pinak08 (commented):
Good job!
Do add the TC and SC for the problem

Pinak08:
Good job!
Do add the TC and SC for the problem

PR #740, opened 2022-07-26
Add files via upload

sundeep1299:


Untitled-1.py
```diff
@@ -0,0 +1,27 @@
+# Time Complexity :o(n)
```
ishika2741, line 1: Kindly check tc

542.py
```diff
@@ -0,0 +1,37 @@
+# Time Complexity : o(n^2)
```
ishika2741, line 1: It would be m*n where m is the no. of rows and n is no. of columns

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1057, opened 2024-02-25
problems solved

kpranavreddy01:


Nandha2Tech (commented):
Great Work! In oranges getting rotten problem when you use dfs TC would become  O(m^2*n^2). For each m*n dfs happens for the entire m*n matrix.

Nandha2Tech:
Great Work! In oranges getting rotten problem when you use dfs TC would become  O(m^2*n^2). For each m*n dfs happens for the entire m*n matrix.

PR #1124, opened 2024-07-30
completed DFS-1

dhikki-study:
######733. Flood Fill#############################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
We have applied BFS by using the direction array where for start box we check the neighbors and color them with new color and further more relates boxes are colored

        

######542. 01 Matrix#########################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : NA

// Your code here along with comments explaining your approach in three sentences only
Here instead of considering 1 we consider 0 as starting point and also converted all 1 to -1 and apply BFS and found the -1 on next level and update as distance

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,80 @@
+######733. Flood Fill####################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work! TC for both the problems would be O(m*n) and SC would also be O(m*n).

Nandha2Tech:
Great Work! TC for both the problems would be O(m*n) and SC would also be O(m*n).

PR #1246, opened 2025-03-02
Completed DFS-1

tejashrijoshi88:


FloodFill.cs
```diff
@@ -0,0 +1,63 @@
+// time complexity O(m*n)
+// space complexity  O(m*n)
+// Approach - we will start from given cell, and check its neighbours, if they are of oroginal color, 
+//mark them with new color. We will go over marking neighbours with new color until we found cell without
+// orginal color
+public class Solution
+{
+    public int[][] FloodFill(int[][] image, int sr, int sc, int color)
+    {
+        Queue<int[]> bfs = new Queue<int[]>();
```
sudhasr, line 10: You have started off with BFS but ended up doing DFS? This Queue is not required. Pls clean your code before pushing it.

sudhasr (commented):
Good solution. Pls do try out the DFS approach for 01Matrix problem. Also, for the FloodFill problem you have started off with BFS but ended up doing DFS.  So Queue is not required, it is an unutilized extra space.

sudhasr:
Good solution. Pls do try out the DFS approach for 01Matrix problem. Also, for the Flood Fill problem you have started off with BFS but ended up doing DFS.  So Queue is not required, it is an unutilized extra space.

PR #1328, opened 2025-07-20
solved

sethu95:


super30admin:
- For `problem1.java`, consider using BFS instead of DFS for better efficiency and to avoid potential stack overflow issues. The DFS function should not return the image at each recursive call; instead, it should modify the image in place and return void.
- For `problem2.java`, the logic for computing distances is incorrect. Consider starting the BFS from all 0 cells simultaneously to compute the distances correctly. The current implementation does not guarantee the shortest distance for all cells.
- Both solutions should include comments to explain the logic and steps, especially for complex parts of the code.
- Ensure that edge cases, such as when the new color is the same as the old color in `problem1.java`, are handled correctly.

PR #1340, opened 2025-08-22
solved 1 questions

nsh2207:


super30admin:
Strengths:
- The student has correctly implemented the BFS approach for both problems.
- The code is generally clean and follows good practices.
- The student has correctly analyzed the time and space complexity.

Areas for Improvement:
- In Question_2.cpp, using a visited array instead of modifying the input matrix would be more robust and maintain the integrity of the input data.
- Adding more comments to explain the logic, especially for the multi-source BFS in Question_2.cpp, would improve readability.
- Avoid using magic numbers like {-1,-1} in Question_2.cpp; instead, use named constants or enums to make the code more understandable.

PR #1370, opened 2025-10-29
done

sejalrj:


super30admin:
Strengths:
- Clear and readable code with appropriate variable naming.
- Good use of Python's `deque` for efficient queue operations.

Areas for Improvement:
- In `updateMatrix`, consider modifying the input matrix directly instead of using an additional `res` matrix to save space.
- In `floodFill`, the condition `if image[new_r][new_c] == color` could be checked once before the loop to avoid redundant checks.
- Adding comments to explain the BFS steps could improve code clarity for others.

PR #1388, opened 2026-01-24
DFS 1 completed

sujaygijre:


super30admin:
Your solution has a good structure and uses BFS, which is a valid approach for flood fill. However, there are several issues that need to be addressed:

1. The most critical issue is that you are not checking the color of adjacent cells before adding them to the queue. You should only add adjacent cells if they have the same original color as the starting pixel. Without this check, your algorithm will try to process cells that should not be changed, leading to incorrect results.

2. The `isVisited` function is confusing because it inserts the cell into the visited set when called. This is not typical for a function that checks if something is visited. It would be clearer to separate the check and the marking as visited. Consider using a separate 2D boolean array for visited status, which is more efficient than using a set of strings.

3. The lambda functions are overused and make the code harder to read. While lambdas can be useful for small, repetitive tasks, having too many can clutter the code. For example, the `getKey` lambda is used only once and might be better inlined.

4. The condition in `addAdjacentCells` calls `isVisited` and only adds the cell if it returns true. However, `isVisited` returns true only when the cell is not visited and then marks it as visited. This is misleading because the function name suggests it only checks. Renaming it to `markVisited` or changing the logic would be better.

5. The reference solution checks the original color at the beginning and returns immediately if the new color is the same as the original. This avoids unnecessary processing. Your solution does not have this check, which could lead to unnecessary work in cases where the color is the same.

6. Consider using a 2D boolean array for visited instead of an unordered_set of strings. This would be more efficient in terms of both time and space.

To improve your solution:
- Add a check for adjacent cells: only process if the cell has the original color.
- Use a 2D visited array instead of a set of strings.
- Remove unnecessary lambdas or simplify them.
- Add the check for same color at the beginning.

